PR #2, opened 2020-10-19
Don't create a context window with width <= 0.

seandewar:
Errors are possible when trying to create a context window for another window that is too small.

Example image of the error

By the way, thanks for creating this plugin; it's very nice! :)


romgrk:
Cool, thanks :)

PR #4, opened 2020-10-20
Fix E11 and improve get_gutter_width() accuracy

seandewar:
Hi again! I thought I'd try and fix 2 other issues that I've seen:

- This PR fixes E11 errors when opening the command-line window while the context window is open (see https://github.com/neovim/neovim/pull/10738):
E11 error image

- This PR also improves the accuracy of `get_gutter_width()` by using Vim's `wincol()` function:
image
This improvement also seems to eliminate the need for my previous PR #2 for small windows.

Cheers!

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -65,33 +65,11 @@ local get_context = function(opts)
 end
 
 local get_gutter_width = function()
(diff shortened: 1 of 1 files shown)


romgrk:
LGTM :)

PR #9, opened 2020-10-21
More improvements.

seandewar:
Hi again!

I've made some more small changes; details are in the commit messages.

The `WinLeave` split bug is a bit strange, though:
split bug

Might possibly be caused by an issue with `nvim_win_set_config()`?
WinEnter is triggered before the file is loaded, but after the window is split when splitting windows, which maybe causes that issue _somehow_, but the same thing does not happen if `nvim_win_open()` is used instead...


romgrk:
Awesome, thanks :)

PR #11, opened 2020-10-23
Revert "Improve get_gutter_width() accuracy using wincol()"

romgrk:
This reverts commit e599c7e2515f3b6be7f3b16eb5e934f1905b5c09.

@seandewar I need to revert this commit due to the bug shown below. Modifying the cursor position mid-function clears the [curswant](https://github.com/neovim/neovim/blob/b5cd052037c532e5416919e4077b721ce6ab3a38/src/nvim/buffer_defs.h#L1140) property. If you have a different proposition to measure the gutter width let me know. Also I'm not sure which issue it solved but if you encounter it please open it so it's documented :)

demo

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -42,11 +42,33 @@ local transform_line = function(line)
 end
 
 local get_gutter_width = function()
-  local old_col = api.nvim_call_function('col', { '.' })
-  api.nvim_call_function('cursor', { 0, 1 })
-  local gutter_width = api.nvim_call_function('wincol', {}) - 1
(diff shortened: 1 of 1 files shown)


seandewar:
Opened PR #12 which should fix this. 👍 

> Also I'm not sure which issue it solved but if you encounter it please open it so it's documented :)

This image from PR #4 shows the issue with the old method:
img

The old method didn't consider `scl` settings such as `yes:X`, `auto:X` and `number`. It was also leaving 2 columns with `auto` set even if there were no signs to display for me.

PR #12, opened 2020-10-23
Fix wincol() approach for get_gutter_width() using getcurpos() and setpos()

seandewar:
Now restores window view state, such as `curswant`.
Should solve the issue in #11.

Let me know if there's any more issues! 👍

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -42,33 +42,13 @@ local transform_line = function(line)
 end
 
 local get_gutter_width = function()
-  local width = api.nvim_win_get_width(0)
-  local number_width = math.max(
-    api.nvim_win_get_option(0, 'numberwidth'),
-    #tostring(api.nvim_call_function('line', { '$' })) + 1
-  )
-  local number = api.nvim_win_get_option(0, 'number')
-  local relative_number = api.nvim_win_get_option(0, 'relativenumber')
-  number_width = (number or relative_number) and number_width or 0
-  local fold_width = api.nvim_win_get_option(0, 'foldcolumn')
+  local saved_cursor = api.nvim_call_function('getcurpos', {})
 
-  local sign_width = 0
+  api.nvim_call_function('cursor', { 0, 1 })
+  local gutter_width = api.nvim_call_function('wincol', {}) - 1
 
-  local sign_column = api.nvim_win_get_option(0, 'signcolumn')
-
-  if sign_column == 'yes' then
-    sign_width = 2
-  elseif sign_column == 'auto' then
-    local signs = api.nvim_call_function('execute', {
-      'sign place buffer=' .. api.nvim_get_current_buf(),
-    })
-    local signs = vim.split(signs, '\n', true)
-    sign_width = #signs > 2 and 2 or 0
-  else
-    sign_width = 0
-  end
-
-  return number_width + fold_width + sign_width
+  api.nvim_call_function('setpos', { '.', saved_cursor })
+  return gutter_width
 end
 
 local nvim_augroup = function(group_name, definitions)
@@ -173,10 +153,6 @@ function M.open()
 
   local gutter_width = get_gutter_width()
   local win_width = api.nvim_win_get_width(0) - gutter_width
-
-  if win_width <= 0 then
-    return
-  end
 
   if winid == nil or not api.nvim_win_is_valid(winid) then
     winid = api.nvim_open_win(bufnr, false, {


romgrk:
Interesting, but I still worry that calling a cursor-movement function in a CursorMoved autocmd is a bit intense, even with the view restore :/

I thought I had found something with `wincol() - col('.')` but then I realized it deosn't work when scrolled.

I've also opened https://github.com/vim/vim/issues/7190.

romgrk:
I FOUND :D 

```
let gutter_width = wincol() - virtcol('.') + winsaveview().leftcol
```

It works right?

seandewar:
`leftcol` represents the left-most column of the buffer being displayed, which is usually only useful when `nowrap` is set and you've horizontally scrolled the view. It's not a virtual column number.

> I still worry that calling a cursor-movement function in a CursorMoved autocmd is a bit intense

I don't believe so. The cursor movement shouldn't trigger a screen redraw or any extra `CursorMoved` events. It's relatively lightweight.

romgrk:
Indeed I see :/

Two last questions:

1. What was wrong with this one?
```vim
augroup gutter
  au!
  au CursorMoved * :echom GutterWidth()
augroup END

function! GutterWidth()
  let numberwidth = max([&numberwidth, strlen(line('$'))+1])
  let numwidth = (&number || &relativenumber)? numberwidth : 0
  let foldwidth = &foldcolumn

  if &signcolumn == 'yes'
    let signwidth = 2
  elseif &signcolumn == 'auto'
    let signs = execute(printf('sign place buffer=%d', bufnr('')))
    let signs = split(signs, "\n")
    let signwidth = len(signs)>2? 2: 0
  else
    let signwidth = 0
  endif
  let gutter_width =  numwidth + foldwidth + signwidth
  return gutter_width
endfunc
```

2. Do you have any opinion on `winsaveview`/`winrestview` vs `getcurpos`/`setcurpos`? Could the `*curpos` ones be more performant?

Sorry to be so picky on this one, it's just that a function that runs on every cursor movement needs to be as efficient as possible, otherwise overtime it accumulates and you end up with a slow cursor because people don't care enough about performance :/ Just let me know your thoughts on the 3 propositions above, I'll do some profiling.

seandewar:
Ah, I forgot about `getcurpos()` with `setpos()`; that method seems to work!

> What was wrong with this one?

This image from PR #4 shows the issue with the old method:
img

That method didn't consider `scl` settings such as `number` and neovim's `yes:X` and `auto:X` settings. It was also leaving 2 columns with `auto` set even if there were no signs to display for me (at least the Lua version was).

Anyway, I wrote a crude benchmark to test each approach out:

```lua
local function winrestview_method()
  local saved_view = vim.api.nvim_call_function('winsaveview', {})

  vim.api.nvim_call_function('cursor', { 0, 1 })
  local gutter_width = vim.api.nvim_call_function('wincol', {}) - 1

  vim.api.nvim_call_function('winrestview', { saved_view })
  return gutter_width
end

local function setpos_method()
  local saved_cursor = vim.api.nvim_call_function('getcurpos', {})

  vim.api.nvim_call_function('cursor', { 0, 1 })
  local gutter_width = vim.api.nvim_call_function('wincol', {}) - 1

  vim.api.nvim_call_function('setpos', { '.', saved_cursor })
  return gutter_width
end

local function old_method()
  local width = vim.api.nvim_win_get_width(0)
  local number_width = math.max(
    vim.api.nvim_win_get_option(0, 'numberwidth'),
    #tostring(vim.api.nvim_call_function('line', { '$' })) + 1
  )
  local number = vim.api.nvim_win_get_option(0, 'number')
  local relative_number = vim.api.nvim_win_get_option(0, 'relativenumber')
  number_width = (number or relative_number) and number_width or 0
  local fold_width = vim.api.nvim_win_get_option(0, 'foldcolumn')

  local sign_width = 0
  local sign_column = vim.api.nvim_win_get_option(0, 'signcolumn')

  if sign_column == 'yes' then
    sign_width = 2
  elseif sign_column == 'auto' then
    local signs = vim.api.nvim_call_function('execute', {
      'sign place buffer=' .. vim.api.nvim_get_current_buf(),
    })
    local signs = vim.split(signs, '\n', true)
    sign_width = #signs > 2 and 2 or 0
  else
    sign_width = 0
  end

  return number_width + fold_width + sign_width
end

function bench(func, name, tries)
  print('benchmarking ' .. name .. ' for ' .. tries .. ' tries')

  local start = vim.api.nvim_call_function('reltime', {})

  for i = 1, tries do
    func()
  end

  local elapsed = vim.api.nvim_call_function('reltime', { start })
  elapsed = vim.api.nvim_call_function('reltimestr', { elapsed })

  print(name .. ' took ' .. elapsed .. ' seconds')
end

function bench_all(tries)
  bench(old_method, 'old method', tries)
  bench(winrestview_method, 'winrestview() method', tries)
  bench(setpos_method, 'setpos() method', tries)
end
```

Here are the results after 100,000 runs of each function while viewing a 10k lines buffer with a few windows open on neovim compiled with Lua JIT:
benchmarks

Each method runs very fast, with the `winrestview()` method being a bit slower as expected. The old and `winrestview()` methods seem to fluctuate a few seconds after every run, but there's seemingly a clear winner with `getcurpos()` and `setpos()`, which always has a consistently low running time!

I've updated this PR to use that method instead. Let me know what you think. 👍

romgrk:
Ok, I think it runs fast enough :smile:

I'll merge as is, thanks for running the benchmarks!

PR #27, opened 2021-01-27
Use nvim-treesitter locals queries

oberblastmeister:
From [issue](https://github.com/romgrk/nvim-treesitter-context/issues/25). This pr allows `nvim-treesitter-context` to use the locals queries provided by `nvim-treesitter`. It also allows for multi line context. I can't get the highlighting to extract the colors from the saved buffer so for now i just call `ts_highlight.attach(bufnr, ft)`which has worse syntax highlighting. Any ideas about how to implement proper syntax highlighting other than calling `ts_highlight.attach`?

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -36,7 +36,11 @@ end
 local get_lines_for_node = function(node)
   local start_row = node:start()
   local end_row   = node:end_()
-  return api.nvim_buf_get_lines(0, start_row, end_row + 1, false)
+  return api.nvim_buf_get_lines(0, start_row, end_row + 1, false)[1]
+end
+
+local function get_lines_for_multiple_nodes(nodes)
+  return vim.tbl_map(get_lines_for_node, nodes)
 end
 
 local get_gutter_width = function()
@@ -95,12 +99,6 @@ function M.get_context(opts)
   return matches
 end
 
-function tbl_reverse(tbl)
-  for i=1, math.floor(#tbl / 2) do
-    tbl[i], tbl[#tbl - i + 1] = tbl[#tbl - i + 1], tbl[i]
-  end
-end
-
 function M.get_parent_matches()
   local contains = vim.tbl_contains
 
@@ -119,8 +117,6 @@ function M.get_parent_matches()
       break
     end
   end
-
-  tbl_reverse(parent_matches)
 
   return parent_matches
 end
@@ -207,7 +203,7 @@ function M.open()
   local gutter_width = get_gutter_width()
   local win_width = api.nvim_win_get_width(0) - gutter_width
 
-  local lines = get_text_for_multiple_nodes(current_node)
+  local lines = get_lines_for_multiple_nodes(current_node)
   if #lines <= 0 then
     return
   end
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -3,6 +3,7 @@ local api = vim.api
 local ts = vim.treesitter
 local Highlighter = ts.highlighter
 local ts_utils = require'nvim-treesitter.ts_utils'
+local ts_highlight = require('nvim-treesitter.highlight')
 local parsers = require'nvim-treesitter.parsers'
 
 -- Script variables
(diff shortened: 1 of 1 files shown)


p00f:
it shows the first `require` 
image

romgrk:
Sorry for the delay.

I haven't read everything yet but I'll test it tomorrow and get back to you soon, but overall it looks good. Two things:
 - How is `CursorMovedVertical` used?
 - Can you resolve the conflict?

Thanks!

romgrk:
Ok, more things from my 1 minute testing:

400ms is slow, can we increase that? I feel like it runs fast enough that the throttling isn't necessary, but let me know what you think.

The context shows things I don't wanna see in context, such as `local api = vim.api` in the screenshot below, from `lua/treesitter-context.lua`.
Screenshot from 2021-02-16 00-44-23
Screenshot from 2021-02-16 00-44-14

Sometimes the context shows more lines over the cursor line, we should probably do something about that? I'd also provide an option to set the maximum context lines an user wants to see.
Screenshot from 2021-02-16 00-45-01
Related: #28

oberblastmeister:
@romgrk, `CursorMovedVertical` is used so that the context only updates when the cursor moves vertically. It is just a custom autocommand that wraps `CursorMoved` and checks if it was vertical. Originally you were just using `CursorMoved`, which I think was the reason why https://github.com/romgrk/nvim-treesitter-context/issues/16 happened, which is fixed with the new autocommand and also prevents updating context unnecessarily which can improve performance. Yes, I will try to fix the merge conflict.

The 400 ms can be decreased, but I don't want this to be slow. The truth is that even at 400 ms it can be slow on large files, which is mainly due to the slowness of `nvim-treesitter` queries, not this plugin :frowning: . In addition the queries are not async so we will probably just have to wait until they get faster.

For the context not showing the correct things that is probably due to the `lua` queries. For other languages like `rust` the context displays perfectly. My idea was to piggy back on the queries provided by `nvim-treesitter`, and for the languages that don't work we can write our own queries. We can put a `queries/` directory in the plugin root and then have files like `queries/lua/context.scm` for our own queries. This will minimize the work that we have to do but at the same time have more accurate context than hardcoded regex.

oberblastmeister:
Also I have made the number of lines configurable. This probably warrants a `setup` function to be able to configure the plugin, what do you think?

romgrk:
I would make both the number of lines and the throttle delay configurable. I've never been very fond of setup functions, but those things have crept their way into the lua plugins so yes I guess we could add a setup function.

For the number of lines, does it show the X last lines or the X first lines? Is that also configurable?

Can you give me examples of large files to test with? I want to make sure the cause is the queries and not us.

For the unwanted lines, would you be able to investigate more? I'd rather not merge an annoying implementation if we can do something about it, in particular if it's for the precise language this plugin is written in. Maybe the lua queries are capturing some sort of Program (root) node that we can remove?

See also the inline lint comments.

lua/treesitter-context.lua
```diff
@@ -61,12 +69,76 @@ local nvim_augroup = function(group_name, definitions)
...
+    return
+  end
+
+  local iter = buf_query:query():iter_captures(target_node, buf, start_row, end_row)
+
+  for capture, node in iter do
+
+    local hl = buf_query.hl_cache[capture]
+
+    local atom_start_row, atom_start_col, atom_end_row, atom_end_col = node:range()
```
romgrk, line 96: No newline between adjacent declarations

lua/treesitter-context.lua
```diff
@@ -61,12 +69,76 @@ local nvim_augroup = function(group_name, definitions)
...
+      api.nvim_buf_set_extmark(bufnr, ns, hl_start_row, hl_start_col, {
+        end_line = hl_end_row,
+        end_col = hl_end_col,
+        hl_group = hl
+      })
+    end
+  end
+end
+
+local function remove_dup(tbl)
```
romgrk, line 118: `remove_duplicate(s?)`, I'd rather spell things out long.

lua/treesitter-context.lua
```diff
@@ -61,12 +69,76 @@ local nvim_augroup = function(group_name, definitions)
...
+  local hash = {}
+  local res = {}
+
+  for _,v in ipairs(tbl) do
+    if (not hash[v]) then
+      res[#res+1] = v
+      hash[v] = true
+    end
+  end
```
romgrk, line 127: Space after comma (`_, v`) and around operators (`#res + 1`)

lua/treesitter-context.lua
```diff
@@ -92,16 +164,40 @@ function M.get_context(opts)
...
 
+function M.get_parent_matches()
+  if not parsers.has_parser() then return nil end
+
+  local contains = vim.tbl_contains
+
+  local matches = ts_query.get_capture_matches(0, '@scope.node', 'locals')
+
+  local curr_node = ts_utils.get_node_at_cursor()
+  if not curr_node then return end
```
romgrk, line 175: This can go at the start of the function no? It avoids calling `ts_query.get_capture_matches(0, '@scope.node', 'locals')`

Also remove newline between adjacent declarations.

romgrk: Oh also `current_node` instead of `curr_node`

lua/treesitter-context.lua
```diff
@@ -139,26 +271,41 @@ function M.close()
...
     return
   end
 
-  if current_node == previous_node then
+  if context_nodes == previous_node then
     return
   end
 
-  previous_node = current_node
+  previous_node = context_nodes
```
romgrk, line 282: Rename `previous_node` to `previous_nodes`

lua/treesitter-context.lua
```diff
@@ -169,83 +316,56 @@ function M.open()
...
-  for _, buf_query in pairs(buf_queries) do
-    if buf_query == nil then
-      break
-    end
-    local iter = buf_query:query():iter_captures(target_node, saved_bufnr, start_row, end_row)
-
-    for capture, node in iter do
-
-      local hl = buf_query.hl_cache[capture]
+  local highlight_better = true
```
romgrk, line 328: Not sure why it's there but it's a constant if condition. Remove it?

romgrk (commented):
Beyond the inline comments, 2 more general comments:
 - You can split utility functions in a separate `treesitter-context.utils` file (or files), it can help readability
 - Regarding the `CursorMovedVertical`, why not just return from the `CursorMoved` handler when `cursor_moved_vertical()` is false? The thing is by dispatching a separate `User` autocommand, you're basically using the autocommands mechanism as a function call, so why not just use a function call?

romgrk:
Did some work in #32 which will conflict with this one, sorry about that. Please rebase if you intend to continue.

romgrk:
The PR is marked as merged because I used it as a base for #36, but I disabled some things I wasn't fully sure of yet, such as the async behavior & the TS queries, which weren't returning exactly the restult I needed. Eg, they would return the `statement_block` (scope node), but I'd rather have its parent node (`if_statement`, `for_statement`, etc). Anyway, if you're still interested in completing this you can restart based on master, it contains most of the code present here, minus some adjustments because TSContext now supports mergin multi-line nodes in a single line, so the highlighting code will diverge.

Issue #29, opened 2021-02-07
How can I change the background color?


weilbith:
Well, there is not much to add to the title. :laughing:

romgrk:
The popup uses the `NormalFloat` group as all other popups. It's not possible to configure this popup individually at the moment, but if it's something you're interested in, it's just a matter of setting `set winhl=Normal:CustomNormal` (or `nvim_win_set_option('winhl', 'Normal:CustomNormal')` in lua). I don't have much time these days but you can open a PR to add the option.

PR #30, opened 2021-02-11
Add highlight group for customization

weilbith:
Fixes: #29

I don't know how to use the API to define highlight links. Thereby it is a plain command string for the moment. I'm glad if someone could help here.

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -175,6 +175,8 @@ function M.open()
     })
   end
 
+  api.nvim_win_set_option(winid, 'winhl', 'NormalFloat:TreesitterContext')
+
   local start_row, start_col = current_node:start()
   local end_row = start_row + 1
   local end_col = 0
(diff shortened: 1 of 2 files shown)


lua/treesitter-context.lua
```diff
@@ -260,6 +262,7 @@ M.enable()
 
 api.nvim_command('command! TSContextEnable  lua require("treesitter-context").enable()')
 api.nvim_command('command! TSContextDisable lua require("treesitter-context").disable()')
+vim.cmd('highlight default link TreesitterContext NormalFloat')
```
romgrk, line 265: One small adjustment here, use `api.nvim_command` like the above instead of `vim.cmd`. Not sure which one is best but I'd rather it be consistent.

weilbith: I pushed a fixup commit for it.

weilbith:
Please don't merge yet. I think something got broken. I'll fix it after work.

romgrk:
Yes, it's `api.nvim_command`, not `vim.nvim_command` :)

weilbith:
Oh dear. :see_no_evil: 
Sorry, I wasn't active over the weekend. I just made the quickfix. I also rebased the branch to include the fixup commits. Should be fine from my point of view.

romgrk:
Ok cool. I haven't tested, but it looks fine, so I'm gonna merge it anyway because I feel lazy today.

Thanks for the PR :)

PR #34, opened 2021-03-11
Close context window on QuitPre

ndreas:
I usually work with multiple windows and multiple tabs, and I discovered that if you have a window with a context window open as the last window in a tab, and you try to close with `:q`, you get an `E5601`:

I added a close autocmd on `QuitPre` which prevents the error.


romgrk:
LGTM, thanks :)

PR #48, opened 2021-05-24
Add option to enable/disable plugin at start

LhKipp:
Hi,

I am still a big fan of this plugin. Thanks man :)

I missed an option to disable this plugin at load. Maybe you want to have it upstream :)


romgrk:
Cool, thanks

PR #54, opened 2021-06-15
Debounce context update for performance improvement

stevearc:
Here's a test of before / after using synchronized tmux panes
https://user-images.githubusercontent.com/506791/122118297-ba7a2b80-cddc-11eb-8d06-d067cac2f7bd.mp4

There was already a method for throttling the update, but it wasn't called in the important places and the logic could be simplified.

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -295,38 +295,21 @@ function M.update_context()
   end
 end
 
-local async_wrap = function(f)
-  local wrapped = function(...)
-    local handle
-    handle = vim.loop.new_async(vim.schedule_wrap(function(...)
-      f(...)
-      handle:close()
-    end))
-    handle:send(...)
-  end
-
-  return wrapped
-end
-
 do
   local running = false
-  local timer
 
   function M.throttled_update_context()
-    if running == false then
-      running = true
+    if running then return end
+    running = true
+    vim.defer_fn(function()
+      local status, err = pcall(M.update_context)
 
-      vim.defer_fn(async_wrap(function()
-        local status, err = pcall(M.update_context)
+      if not status then
+        print('Failed to get context: ' .. err)
+      end
 
-        if not status then
-          print('Failed to get context: ' .. err)
-        end
-
-        running = false
-        if timer then timer:close() end
-      end), 400)
-    end
+      running = false
+    end, 100)
   end
 end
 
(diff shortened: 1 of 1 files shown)


lua/treesitter-context.lua
```diff
@@ -295,38 +295,21 @@ function M.update_context()
...
-    handle = vim.loop.new_async(vim.schedule_wrap(function(...)
-      f(...)
-      handle:close()
-    end))
-    handle:send(...)
-  end
-
-  return wrapped
-end
-
```
stevearc, line 310: We shouldn't need any of this because `vim.defer_fn` automatically wraps the callback. See https://neovim.io/doc/user/lua.html#vim.defer_fn

romgrk:
Thanks

romgrk:
Sorry had to make it into an option because it was causing issues on my setup that I don't have time to debug.

```lua
require'treesitter-context'.setup{
    enable = true, -- Enable this plugin (Can be enabled/disabled later via commands)
    throttle = true, -- Throttles plugin updates (may improve performance)
}
```

stevearc:
Oh damn! I'll dig in and see if I can find what might be going wrong. What are the problems that you're seeing?

gegoune:
Would be nice to be able to configure it within `require('nvim-treesitter.configs').setup({})` table.

romgrk:
When the context window is opened, the current buffer isn't marked anymore as current in [barbar](https://github.com/romgrk/barbar.nvim), so there are probably autocommands that are running but shouldn't be running when the context window is opened. We should use `noautocmd` because there is really no reason for other plugins to listen to the state of our window & buffer.

stevearc:
Tested adding `noautocmd = true` to the `nvim_open_win`, and it does fix the focus issue with barbar. Makes sense; opening the window from a `defer_fn` is going to allow all the normal autocmds to fire, where previously the window was getting opened from inside a non-nested `WinScrolled` autocmd or similar, which would prevent them from firing.

Also, FWIW I'm not seeing the same terrible lag when I scroll through files even without the debouncing code. Maybe something in nvim-treesitter changed to improve performance, or maybe I changed something else in my vim config that put me under the frame budget. Either way, I wouldn't be sad if the debouncing logic got cut to simplify things.

gegoune:
For me this PR made a bit difference. I had context plugin disabled for a while waiting for this PR to be merged. Works really nice now. Thanks @stevearc!

romgrk:
@stevearc If you have enough time to push either a fix or a cleanup go ahead, otherwise I'll just leave as it is until I have time to do stuff.

Issue #66, opened 2021-09-11
Option to limit the height of the floating window


zeertzjq:
In a C `if () ... else if () ... else if () ... else if () ... else ...` block with many branches, the floating window becomes very large, occupying over a half of the screen. An option to limit the height of the floating window can be helpful in this case.

![Screenshot_20210911_154406](https://user-images.githubusercontent.com/35768171/132940603-29e1c52f-c276-4e7d-9135-1bd85269a9a4.png)

jakelinnzy:
Hey, what do you think about #67?

ldelossa:
I also want this but maybe for a bit of a different reason. 

I want to use this plugin *only* to show which function I am inside. Having it show me if I'm in an "if" or a "switch" case is not very helpful to me. Would your pull @jakelinnzy accomplish this by setting the length to 1?

zeertzjq:
Now things are getting a bit complicated, as it is not clear what to truncate.

zeertzjq:
> I also want this but maybe for a bit of a different reason.
> 
> I want to use this plugin _only_ to show which function I am inside. Having it show me if I'm in an "if" or a "switch" case is not very helpful to me. Would your pull @jakelinnzy accomplish this by setting the length to 1?

I think it wouldn't.

romgrk:
Closed in #67, the patterns use-case is done in d4f0193. See readme for details.

PR #67, opened 2021-09-15
Add option 'max_lines'

jakelinnzy:
This should address #66.

Adds option `max_lines` to limit the height of the context window. 0 or negative means no limit, which is the default.

Before:

Screen Shot 2021-09-16 at 2 36 21 am

After: (`max_lines = 1`)

Screen Shot 2021-09-16 at 2 35 54 am


zeertzjq:
Maybe this can also be passed to `get_parent_matches` to improve performance?

jakelinnzy:
@zeertzjq How about this?

romgrk:
Cool, thanks :)

Issue #69, opened 2021-10-02
Sometimes context is shown even that `enable = false`


lyokha:
I temporarily disabled the plugin, but still see the context in C files.

My config is

```lua
require'treesitter-context'.setup{
    enable = false,
    throttle = true,
    max_lines = 0,
    patterns = {
        -- For all filetypes
        -- Note that setting an entry here replaces all other patterns for this entry.
        -- By setting the 'default' entry below, you can control which nodes you want to
        -- appear in the context window.
        default = {
            'class',
            'function',
            'method',
            -- 'for', -- These won't appear in the context
            -- 'while',
            -- 'if',
            -- 'switch',
            -- 'case',
        },
        -- Example for a specific filetype.
        -- If a pattern is missing, *open a PR* so everyone can benefit.
        --   rust = {
        --       'impl_item',
        --   },
    },
}
```

After digging into the code I found that this is because `M.setup()` is called from within *treesitter-context.lua* unconditionally (for initializing `config` if I get it right), but this leads to unconditional call to `M.enable() -> M.throttled_update_context()` which is only good for toggling the plugin in run-time.

Here is a small patch that works well for me.

```diff
diff --git a/lua/treesitter-context.lua b/lua/treesitter-context.lua
index c64228e..3949c5c 100644
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -496,7 +496,7 @@ function M.open()
   end
 end
 
-function M.enable()
+function M.enable(update)
   local throttle = config.throttle and 'throttled_' or ''
   nvim_augroup('treesitter_context', {
     {'WinScrolled', '*',                   'silent lua require("treesitter-context").' .. throttle .. 'update_context()'},
@@ -510,7 +510,10 @@ function M.enable()
     {'User',        'SessionSavePost',     'silent lua require("treesitter-context").open()'},
   })
 
-  M.throttled_update_context()
+  if update then
+    M.throttled_update_context()
+  end
+
   enabled = true
 end
 
@@ -525,7 +528,7 @@ function M.toggleEnabled()
     if enabled then
         M.disable()
     else
-        M.enable()
+        M.enable(true)
     end
 end
 
@@ -551,7 +554,7 @@ end
 
 M.setup()
 
-api.nvim_command('command! TSContextEnable  lua require("treesitter-context").enable()')
+api.nvim_command('command! TSContextEnable  lua require("treesitter-context").enable(true)')
[...]

PR #70, opened 2021-10-02
Avoid direct call to M.throttled_update_context() from M.setup()

lyokha:
See details in #69

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -496,7 +496,7 @@ function M.open()
   end
 end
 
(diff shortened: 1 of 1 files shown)


romgrk:
Sorry, after reviewing the PR I felt like this fix wasn't adressing the root issue. I've added a fix to avoid calling `.enable()` if `.setup({ enabled = false })` is called by the user. Let me know if the issue still appears for you.

lyokha:
I see, your fix looks very sound, it works well, thanks!

PR #71, opened 2021-10-04
Fixed multiline nodes (see #65)

### LINKED ISSUE #65: Multi-line nodes are not working
Screenshot_20210828_054743
Screenshot_20210828_054756

Are there any solutions for this?

lyokha:
Note that the upper node gets matched against values from `config.patterns.default` using `get_type_pattern()` which returns a regexp (i.e. the original value, say `function` gets returned as `%f[%w]function%f[^%w]`  having been wrapped with `word_pattern()`): this means that key `['function']` in `last_types` must be wrapped this same way (as `[word_pattern('function')]`) because its finally gets retrieved with `last_types[type]`.

Note also that `function_declaration` may be deeply nested inside `function_definition` (at least, in C), and therefore a correct implementation requires a recursive traversal of nodes which is implemented with function `find_node()` introduced in this commit.

Closes #65

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -23,7 +23,7 @@ ffi.cdef'int curwin_col_off(void);'
 -- Tells us at which node type to stop when highlighting a multi-line
 -- node. If not specified, the highlighting stops after the first line.
 local last_types = {
-  ['function'] = {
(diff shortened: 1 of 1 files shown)


romgrk:
Cool, thanks :)

PR #81, opened 2021-11-20
Ensure treesitter-context buffer exists

b0o:
Improves compatibility with other plugins such as session managers that like to delete buffers. Previously, if another plugin deleted this plugin's buffer, treesitter-context would error out upon trying to show the context window. Now, the buffer is re-created if it doesn't exist. The buffer is also now deleted when `disable()` is called.

Side note:

<details>
  <summary>Click to expand</summary>

This isn't directly related, but I don't really like having so many module-level state variables. I think the code would be nicer if they were all kept inside a `state` table:

```lua
-- Current
local didSetup = false
local enabled = nil
local winid = nil
local bufnr = api.nvim_create_buf(false, true)
local _bufnr = nil -- Don't access directly, use get_buf()
local ns = api.nvim_create_namespace('nvim-treesitter-context')
local context_nodes = {}
local context_types = {}
local previous_nodes = nil

-- Proposal
local state = {
  didSetup = false,
  enabled = nil,
  winid = nil,
  bufnr = nil, -- Don't access directly, use get_buf()
  ns = api.nvim_create_namespace('nvim-treesitter-context'),
  context_nodes = {},
  context_types = {},
  previous_nodes = nil,
}
```

Then there's no worry about shadowing, which is why I originally changed `bufnr` to `_bufnr`.

</details>

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -61,7 +61,7 @@ local INDENT_PATTERN = '^%s+'
 local didSetup = false
 local enabled = nil
 local winid = nil
-local bufnr = api.nvim_create_buf(false, true)
+local _bufnr = nil -- Don't access directly, use get_buf()
 local ns = api.nvim_create_namespace('nvim-treesitter-context')
 local context_nodes = {}
 local context_types = {}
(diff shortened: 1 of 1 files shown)


romgrk:
Thanks! I agree with your side note, I might change when I have time to or PR welcome.

PR #83, opened 2021-12-04
Use getwininfo() instead of FFI to get gutter width in Nvim 0.6

zeertzjq:
It is no longer necessary to rely on an Nvim internal C function.


romgrk:
Thanks :)

PR #84, opened 2021-12-08
Enable full lua patterns as node types on user demand

FoamScience:
I was trying to make this neat piece of software work with my own TreeSitter grammar when I discovered
that matched node types are mapped with `word_pattern` function. In my grammar, I have `dict` and `dict_body` nodes and
I would like to match only `dict`.

So, I added a user setting to turn off that mapping, and use bare Lua patterns instead. Now I can live with `^dict$`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,12 @@ require'treesitter-context'.setup{
         --       'impl_item',
         --   },
     },
+    exact_patterns = {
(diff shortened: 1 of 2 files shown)


FoamScience:
@romgrk  Still maintaining this thing?

romgrk:
LGTM, sorry for the delay I haven't had time for my open-source projects.

PR #85, opened 2021-12-17
Omit PHP attributes from class/method context.

sebcode:
Bug: The context line shows PHP 8 class or method attributes instead of the class/method definition. This PR is an attempt to fix that.

Before:
0before

With this PR:
1after

:TSPlayground
3tsplayground

Test code:
```php
<?php

use Attribute;

#[Attribute(Attribute::TARGET_CLASS)]
class ClassAttr { }

#[Attribute(Attribute::TARGET_METHOD)]
class MethodAttr { }

/**
 * hello
 */
#[ClassAttr]
class Inventory
{
	/**
	 * @throws \Exception
	 */
	#[MethodAttr]
	public static function testasdf(array $args): void
	{
		// asdf
	}
}
```

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -32,6 +32,17 @@ local last_types = {
   },
 }
 
+-- Tells us which leading child node type to skip when highlighting a
+-- multi-line node.
+local skip_leading_types = {
+  [word_pattern('class')] = {
+    php = 'attribute_list',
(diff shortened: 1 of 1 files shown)


romgrk:
This should be mitigated by fixing the multi-line nodes issue, but I'll merge as is because I currently don't have enough time to fix it.

PR #88, opened 2021-12-19
fix: Packer events with Stablize.nvim error

axieax:
This resolves the error with [stabilize.nvim issue 13](https://github.com/luukvbaal/stabilize.nvim/issues/13) which occurs when trying to run `PackerCompile` with the context window open.

image


axieax:
UPDATE: this does not fix it unfortunately :(

PR #89, opened 2021-12-21
Add option line_numbers to show line numbers of the context items

### LINKED ISSUE #37: Show line number
https://raw.githubusercontent.com/wellle/images/master/context-example.png

Gutawer:
Closes #37 if merged.

This implements the showing of line numbers by adding a second floating window that gets drawn on top of the gutter. A new highlight group has been added so that this can be themed (for example to highlight the different line number) - by default it just links to `LineNr` so it looks like any other line number.

Example with `highlight TreesitterContextLineNumber guifg=#c3e88d`:
image

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -411,7 +411,7 @@ function M.close()
 
   if gutter_winid ~= nil and api.nvim_win_is_valid(gutter_winid) then
     -- Can't close other windows when the command-line window is open
-    if api.nvim_call_function("getcmdwintype", {}) ~= "" then
+    if api.nvim_call_function('getcmdwintype', {}) ~= '' then
       return
     end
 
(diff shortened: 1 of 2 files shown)


lua/treesitter-context.lua
```diff
@@ -577,6 +630,7 @@ api.nvim_command('command! TSContextDisable lua require("treesitter-context").di
 api.nvim_command('command! TSContextToggle  lua require("treesitter-context").toggleEnabled()')
 
 api.nvim_command('highlight default link TreesitterContext NormalFloat')
+api.nvim_command "highlight default link TreesitterContextLineNumber LineNr"
```
romgrk, line 633: One thing, could you use a style consistent with the rest? E.g. here change double quotes for single quotes and add parenthesis, and above in other diffs use single quotes. Thanks :)

Gutawer: Done, sorry about that! Think my auto-formatter got run by accident on some lines.

camgraff:
This patch currently doesn't take `relativenumber` into account. Could that be added?

lf-:
> This patch currently doesn't take `relativenumber` into account. Could that be added?

Note that there is a third mode of numbering where `relativenumber` and `number` are both set, in which it should probably show absolute numbers. See vim help entry `number_relativenumber`.

binhtran432k:
While this PR is waiting to be merged, I have folked it for my personal use. Here is my folk:
https://github.com/binhtran432k/nvim-treesitter-context
It is added auto skip scrolling context and refactor context to only use one floating window for line number
Furthermore, I'm not an expert so I need some author's reference to reimplement this feature. Or should I make a pull request, do anybody have ideas?
Demo:
Recording 2022-02-27 at 19 25 54

lewis6991:
Thanks for this!

I've applied it to my fork too which I'm maintaining: https://github.com/lewis6991/nvim-treesitter-context

lewis6991:
Superseded with #115

romgrk:
@Gutawer Thanks for the PR :) Sorry for the delay

PR #98, opened 2022-02-23
Only open on windows that are taller than some minimum window height

lf-:
This is helpful if you have an editor with a bunch of tiny windows to not have them eaten up by context that there's not space for.


chancez:
Oh I was just looking at how to implement this outside of the plugin, having this would be a huge improvement for me.

lf-:
@lewis6991  can you review? I've rebased against the latest master.

PR #108, opened 2022-04-03
fix #28: cursor is hidden behind context window

### LINKED ISSUE #28: Cursor is hidden behing context window

mg979:
Not sure if `scrolloff` should be adjusted less often, depending on how often that function is actually called.

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -447,11 +447,13 @@ function M.open()
   display_window(win_width, win_height, 0, gutter_width)
 
   -- Adjust &scrolloff to prevent the cursor from moving below the window
-  local so = vim.fn.getwinvar(0, '&scrolloff')
-  if win_height > so then
-      vim.fn.setwinvar(0, '&scrolloff', win_height)
-  else
-      vim.fn.setwinvar(0, '&scrolloff', -1)
+  if config.max_lines <= 0 then
+      local so = vim.fn.getwinvar(0, '&scrolloff')
+      if win_height > so then
+          vim.fn.setwinvar(0, '&scrolloff', win_height)
+      else
+          vim.fn.setwinvar(0, '&scrolloff', -1)
+      end
   end
 
   -- Set text
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -446,6 +446,14 @@ function M.open()
 
   display_window(win_width, win_height, 0, gutter_width)
 
+  -- Adjust &scrolloff to prevent the cursor from moving below the window
+  local so = vim.fn.getwinvar(0, '&scrolloff')
+  if win_height > so then
+      vim.fn.setwinvar(0, '&scrolloff', win_height)
+  else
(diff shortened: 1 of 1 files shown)


mg979:
Not that I'm not using `vim.o.scrolloff` because it's broken in current nvim stable.

mg979:
You can skip the last commit if you think it's not necessary.

romgrk:
Hey @mg979, sorry for the delay I've been AFK for an extended period of time :| I'm not sure what the PR is solving/improving but if you want to re-open it go ahead, either I or @lewis6991 should be around to respond.

mg979:
I thought this repo wasn't maintained anymore. The fix is for when the cursor moves below the popup window. @lewis6991 fixed it in a [different way](9a4ecfa65c5c7567f7bc2a7d505e06115a7f2d2a) in his fork, but I prefer [my fix ](acdcfb21dd7254b481fdd8e77c43f5b1d1edcd6e)because it doesn't truncate the popup lines.

I've been using this
```
  -- Adjust &scrolloff to prevent the cursor from moving below the window
  local so = vim.fn.getwinvar(0, '&scrolloff')
  if win_height > so then
      vim.cmd('noau setlocal scrolloff=' .. win_height)
  else
      vim.cmd('noau setlocal scrolloff=-1')
  end
```

instead of the one proposed in this PR, using `noauto` to prevent problems. I'm using my fork atm so I won't make another PR, maybe you'd prefer @lewis6991 fix anyway.

romgrk:
We've merged @lewis6991 fork in this one and will be transferring it to the nvim-treesitter org shortly, so it will become the official fork and will be under the responsability of more than one person (hehe sorry again for the delays).

I can understand why your fix is interesting, @lewis6991 do you have an opinion?
Maybe I'd integrate this change under an option, as the behavior sounds a bit more invasive than the fix we have merged.
I think it should also include logic to restore the user's `scrolloff` option instead of resetting it to `-1`.

mg979:
>Maybe I'd integrate this change under an option, as the behavior sounds a bit more invasive than the fix we have merged.

I don't really care if my point of view is considered or not, since I settled on using my fork anyway. But I don't think it's more invasive, at least if one uses `noau`. I experimented (briefly) with other solutions, adjusting the viewport to make room for the popup would trigger again `WinScrolled` and other autocommands, and it felt clunky and probably very bad for performance. This seems very smooh and has no drawbacks that I can see.

>I think it should also include logic to restore the user's scrolloff option instead of resetting it to -1.

That's exactly what `-1` does, if you read `:h scrolloff`. At least when using `setlocal` (otherwise an invalid value for scrolloff). Unless you mean a user winlocal scrolloff, yeah this would reset it to the global value, but I don't care.

romgrk:
Alright, since everyone seems happy with their solution let's close again.

PR #125, opened 2022-06-09
feat: add config.mode

Saecki:
Rebased #120  onto master


romgrk:
Let's ping the author of #120, I've added my review there, we can get it merged soon. I'll reopen this if the other PR is not reactive.

PR #128, opened 2022-06-11
fix: show context line if it is covered by popup

### LINKED ISSUE #126: Issue #80 was introduced again: context won't be shown if the line is under context window
After this commit https://github.com/nvim-treesitter/nvim-treesitter-context/commit/81116dc03abf35ca99c7910f9f779961042dea12, the issue https://github.com/nvim-treesitter/nvim-treesitter-context/issues/80 occurs again.

The issue is: the context of a line isn't appearing when the line is covered by the context floating window, and instead it appears only when the line is no longer in the window. This is the previous bug and had been fixed by the old commit (https://github.com/nvim-treesitter/nvim-treesitter-context/commit/c8b09b90d98c336478d2e212c865d42b0dd904bb). However, this commit brought it back again. 

Thank you very much.

Saecki:
Fixes #126, by iterating in reverse order.


lua/treesitter-context.lua
```diff
@@ -336,20 +336,22 @@ local function get_parent_matches(max_lines)
   -- save nodes in a table to iterate from top to bottom
   local parents = {}
   while node ~= nil do
-    parents[#parents+1] = node
+    table.insert(parents, node)
```
lewis6991, line 339: This change is unnecessary.

Saecki: Yeah, ocd kicking in

Saecki:
This allows for configuring whether inner or outer lines are truncated from the context (#121 and #126). A suggestion for a better name for the config option would be welcome, feels a bit weird.

rockyzhang24:
Hi @Saecki It looks good. Another issue is that when we set `mode = ‘topline’`, the same issue happens, i.e., context of the topline won’t be shown when it is covered by the floating context window, and instead only be shown when it  goes out of the window. If you have time you can fix it i this PR. If not, I will open a PR to fix it later today or tomorrow. 

Thank you!

Saecki:
This is a little bit of a hack, to select another child node, for functions most likely some `parameters` or a `return_type`, so that when searching outwards for context, the function node comes up. But for node types that don't have another child node in the first line this won't work.

rockyzhang24:
Based on my fully test, when we set `truncate_side = 'inner'`, your code is very buggy. To reproduce, you can enable all the patterns shown below in nvim-treesitter-context config. 

```lua
  patterns = {
    default = {
      'class',
      'function',
      'method',
      'for',
      'while',
      'if',
      'switch',
      'case',
    },
  },
```

And then write a piece of code with kinda deep nested scope. Then using <C-y> and <C-e> to move the screen. You will notice the bug.

Saecki:
Could you go into more detail what "the bug" your experiencing actually is? And maybe provide the file so I can try to reproduce it? Everything seems to work fine on my side.

rockyzhang24:
As I said, first enable all the patterns in nvim-treesitter-context config. Then write a piece of code with deep nested scopes such as

```
function xxx()
  if xxx then
    if xxx then
      if xxx then
        if xxx then
        end
      end
    end
  end
end
```

Next, adjust the these options. Set `truncate_side = 'inner'` along with `max_lines = 1` and `max_lines = 2`. You will observe the bugs.

Also, set `mode = 'topline` along with `max_lines` set to different values. Then we can see the bugs about the topline are still there. The ideal result should be: when a line is covered by the context floating window, its context should be show. And these newly-shown context will cover more lines and it leads the contexts of these lines should be displayed. 

Thanks.

Saecki:
This is what it looks like on my machine:

[asciicast](https://asciinema.org/a/NVaEiD5pdjWmSNHE4Rcs01YNB)

I would consider that the expected behavior, if you want to see the entire context just set `max_lines` to `0`.

rockyzhang24:
Yes. outer with any max_lines works well. But for inner, it works well only when max_lines is set to 0. You can play around that example with different max_lines and put the cursor at different position. you will find many bugs. 

Thank you.

Saecki:
Same goes for outer with `max_lines` set to `2`:
[asciicast](https://asciinema.org/a/lhXQsk0XyhIgRDp8d1WtFblQG)

What behavior do you expect?

rockyzhang24:
For outer, it works perfectly. But for inner, it not. I think you don't understand the logic. Oh by the way, in you code, you actually opposite the inner and outer cases (look my comment in your commit). So here I said outer probably it's inner in you code.

Anyway, I will try to implement and solve all the issues when I am free.

Saecki:
It does what it says, it truncates (cuts off) the `inner` or `outer` context, as I've said I'm open to name suggestions for the config option. I think your comment is gone after I rebased onto master.

romgrk:
@rockyzhang24 Please include runnable example and screenshots when you report an issue, "you'll find many bugs" is a bit vague and can be deceiving (aka "works on my machine"), Thanks :)

Saecki:
We're trying our best here to fix the issues with that commit. As you've stated above "the bugs" that you're observing (which I still don't really understand) are only happening in `topline` mode, which was only introduced after the offending commit. So this seems to fix the original issue you're having.
Please just share a screenshot or a screen capture that highlights the bugs that are remaining so we can sort them out.

romgrk:
@Saecki as we don't have actionnable bugs to work on, let us know when this is ready for review.

@rockyzhang24 PR welcome.

Saecki:
I would say this is good to go

romgrk:
Two things:
 - Use single-quotes everywhere, as the rest of the project uses that convention
 - There might be an error if there is a typo in `truncate_side`. Maybe handle it this way?:

```lua
if truncate_side === 'inner' then
  -- ...
else
  -- truncate_side === 'outer' case
  -- ...
end
```

Saecki:
That should be it. Also, please tell me if you can think of a nicer name for the option.

rockyzhang24:
I will report it again (the third time and you guys kept ignoring me), you should change your code to 

```lua
if config.truncate_side == 'outer' then  -- default to 'outer'
     return vim.list_slice(
       parent_matches,
       1,
       math.min(#parent_matches, max_lines)
     )
   else  -- 'inner'
     return vim.list_slice(
       parent_matches,
       math.max(1, #parent_matches - max_lines + 1),
       #parent_matches
     )
   end
```
i.e., based on your code, the first case should be outer, and the seconde (`else`) case is inner.

Saecki:
> It does what it says, it truncates (cuts off) the `inner` or `outer` context, as I've said I'm open to name suggestions for the config option.
The option specifies the side from which lines are truncated, so `outer` takes the first `max_lines` lines and `inner` the last ones.

rockyzhang24:
Okay my bad. I thought it meant what side you want to display. So actually it means what side you want to discard. Got it.

rockyzhang24:
Also I think "scope" (or "contexts") can be a better name of this option. `scope = 'inner'` means we want to keep the inner scopes (i.e., truncate the outer), and `scope = 'outer'` means we want to keep the outer scopes. This is more intuitive.

romgrk:
I've also had hesitations about the name of the option, I don't like `truncate_side` much, but TBH not enough to prevent this from being merged.

If you guys have a better name to propose I'll let you settle on something, we'll merge this tomorrow with the name that's there.

Saecki:
Maybe `truncate_scope` or `keep_scope`

Saecki:
I've landed on `trim_scope`, while not perfect, I think it's definitely more fitting.

PR #131, opened 2022-06-14
add object to scala

stumash:
there are nodes like `object_definition` in the scala treesitter, because you can do things like

```scala
object Foo {
  def dank(): Unit = println("foodank")
}
Foo.dank()
```

where in other languages, like java, you would do

```java
class Foo {
  public static void dank() { System.out.println("foodank"); }
}
Foo.dank();

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -99,6 +99,9 @@ local DEFAULT_TYPE_PATTERNS = {
   rust = {
     'impl_item',
   },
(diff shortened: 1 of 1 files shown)


romgrk:
Is there a way to restrict the pattern a bit more? What node types is `object` going to match? If only `object_definition` is the target, it's better to leave it at that.

romgrk:
thanks :)

PR #139, opened 2022-06-29
fix: set option nowrap on context window

### LINKED ISSUE #138: Outer-most context linewraps over inner contexts
If I have a function:
```c
PetscErrorCode CreateOperatorForDomain(Ceed ceed, DM dm, SimpleBC bc,
                                       CeedData ceed_data, Physics phys,
                                       CeedOperator op_apply_vol,
                                       CeedOperator op_apply_ijacobian_vol,
                                       CeedInt height,
                                       CeedInt P_sur, CeedInt Q_sur,
                                       CeedInt q_data_size_sur, CeedInt jac_data_size_sur,
                                       CeedOperator *op_apply, CeedOperator *op_apply_ijacobian) {
//[..........]
  if (phys->has_neumann || 1) {
    // --- Setup
    ierr = DMGetLabel(dm, "Face Sets", &domain_label); CHKERRQ(ierr);

    // --- Get number of quadrature points for the boundaries
    CeedInt num_qpts_sur;
    CeedBasisGetNumQuadraturePoints(ceed_data->basis_q_sur, &num_qpts_sur);

    // --- Create Sub-Operator for inflow boundaries
    for (CeedInt i=0; i < bc->num_inflow; i++) {
      CeedVector          q_data_sur, jac_data_sur;
      CeedOperator        op_setup_sur, op_apply_inflow,
                          op_apply_inflow_jacobian = NULL;
      CeedElemRestriction elem_restr_x_sur, elem_restr_q_sur, elem_restr_qd_i_sur,
                          elem_restr_jd_i_sur;
```
it'll show the context as:
image
So the function is wrapped into one line, but then cutoff because it's too long. Perfectly fine.

If I move past the `if` statement, I see:
image
So the excess of the function signature context has now been wrapped onto the next context line. Note that the line number for the `if` context is correct, but it's not showing the correct one. Note that the function signature still isn't complete...

Continuing to move past the `for` loop, I see:
image
So now the function signature been wrapped over-top of both the `if` and `for` statements. Not ideal.

If I expand the width of the screen such that the function signature context can fit on two lines, I get:
image
Where the `if` context is now there and the line number next to it is incorrect (that's for the `for` context).

This seems to be an intended consequence of the patch that closes #10. Or perhaps an overlooked issue when making the patch for that.

I think the less annoying solution is just to keep each context limiited to it's own line. Being able to see the full context line would be nice, but not in exchange of not being able to see the other contexts.

romgrk:
Closes #138 

Explicitly set the `nowrap` option on the context window.

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -333,8 +333,9 @@ local function display_window(bufnr, winid, width, height, col, ty, hl)
       border = sep and {'', '', '', '', sep, sep, sep, ''} or nil,
     })
     api.nvim_win_set_var(winid, ty, true)
-    vim.wo[winid].winhl = 'NormalFloat:'..hl
-    vim.wo[winid].foldenable = false
+    api.nvim_win_set_option(winid, 'wrap', false)
+    api.nvim_win_set_option(winid, 'foldenable', false)
+    api.nvim_win_set_option(winid, 'winhl', 'NormalFloat:' .. hl)
   else
     api.nvim_win_set_config(winid, {
       win = api.nvim_get_current_win(),


jrwrigh:
Anything preventing this from being merged?

romgrk:
@jrwrigh just to confirm, this PR solves #138?

jrwrigh:
Yep!

romgrk:
Alright, let's merge then. I'll leave the `nvim_win_set_option` as it's locally consistent with `nvim_win_set_var`.

PR #142, opened 2022-07-02
feat: add markdown pattern

jghauser:
This adds a pattern to show the current section in the markdown file. It works with headings and subheadings also. This uses the new markdown parser that was recently added to nvim-treesitter.


romgrk:
Thanks :)

PR #156, opened 2022-09-02
Add terraform default type patterns

rwblokzijl:
Terraform uses `block`s for all their top level data structures. Ideally I'd add a query for those that have an identifier that starts with `resource`, `data`, `locals` or `provider`. But that doesn't seem possible within this plugin.


fourstepper:
I've tested this right now and it works well.

PR #162, opened 2022-09-07
fix: inherit tabstop on context window

Bahex:
Addresses #146

Sets `tabstop` same as the current buffer.


younger-1:
Tested. Works fine in local machine

PR #170, opened 2022-10-03
feat: Add side scrolling to context

rwblokzijl:
This PR aligns the side scroll of the context to the side scroll of the page.

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -625,13 +625,12 @@ local function horizontal_scroll_contexts()
   if context_winid == nil then
     return
   end
-  local current_win_view = vim.api.nvim_win_call(context_winid, function ()
-    return vim.fn.winsaveview()
-  end)
-  if vim.fn.winsaveview()["leftcol"] ~= current_win_view["leftcol"] then
-    current_win_view["leftcol"] = vim.fn.winsaveview()["leftcol"]
+  local active_win_view = vim.fn.winsaveview()
+  local context_win_view = vim.api.nvim_win_call(context_winid, function () return vim.fn.winsaveview() end)
+  if active_win_view["leftcol"] ~= context_win_view["leftcol"] then
+    context_win_view["leftcol"] = active_win_view["leftcol"]
     vim.api.nvim_win_call(context_winid, function ()
-      return vim.fn.winrestview(current_win_view)
+      return vim.fn.winrestview(context_win_view)
     end)
   end
 end
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -621,6 +621,15 @@ local function get_relative_line_num(ctx_node_line_num)
   return cursor_line_num - ctx_node_line_num - num_folded_lines
 end
 
+local function side_scroll_contexts()
+  local current_win_view = vim.api.nvim_win_call(context_winid, function ()
+    return vim.fn.winsaveview()
+  end)
+  current_win_view["leftcol"] = vim.fn.winsaveview()["leftcol"]
+  vim.api.nvim_win_call(context_winid, function ()
+    return vim.fn.winrestview(current_win_view)
+  end)
+end
 
 local function open(ctx_nodes)
   local bufnr = api.nvim_get_current_buf()
@@ -671,6 +680,7 @@ local function open(ctx_nodes)
     table.insert(lno_text, build_lno_str(line_num, gutter_width-1))
   end
 
+  side_scroll_contexts()
   set_lines(gbufnr, lno_text)
   if not set_lines(ctx_bufnr, context_text) then
     -- Context didn't change, can return here
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -315,6 +315,18 @@ local function get_gutter_width()
   return vim.fn.getwininfo(vim.api.nvim_get_current_win())[1].textoff
 end
 
+local window_scrolled_horizontal
+do
+  local leftcol
+  window_scrolled_horizontal = function()
+    local newcol = vim.fn.col('.') - vim.fn.wincol() + get_gutter_width()
+    if newcol ~= leftcol then
+      leftcol = newcol
+      return true
+    end
+    return false
+  end
+end
 local cursor_moved_vertical
 do
   local line
@@ -621,7 +633,10 @@ local function get_relative_line_num(ctx_node_line_num)
   return cursor_line_num - ctx_node_line_num - num_folded_lines
 end
 
-local function side_scroll_contexts()
+local function horizontal_scroll_contexts()
+  if context_winid == nil then
+    return
+  end
   local current_win_view = vim.api.nvim_win_call(context_winid, function ()
     return vim.fn.winsaveview()
   end)
@@ -680,7 +695,6 @@ local function open(ctx_nodes)
     table.insert(lno_text, build_lno_str(line_num, gutter_width-1))
   end
 
-  side_scroll_contexts()
   set_lines(gbufnr, lno_text)
   if not set_lines(ctx_bufnr, context_text) then
     -- Context didn't change, can return here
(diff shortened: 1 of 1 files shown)


rwblokzijl:
Marked as draft due to performance issues. Input welcome

rwblokzijl:
Take the following CSV in terraform:

Scrolling down with context it looks like:

Before the PR, if you scroll to the right, the context doesn't change, thus misaligning the CSV header:

With this PR the side scroll of the context window will follow the side scroll of the active window, thus aligning the columns properly:

lewis6991:
`winsaveview` and `winrestview` are very expensive. You should avoid calling them as much as possible.

rwblokzijl:
I was able to eliminate 1 `winsaveview`, however i can't find a way of scrolling a non-active window without them. If you have any ideas I'll look into them.

Regarding performance i have to issues and everything scrolls smoothly. The earlier performance issue i mentioned on this PR was due to another plugin slowing down horizontal scrolling.

lua/treesitter-context.lua
```diff
@@ -671,6 +680,7 @@ local function open(ctx_nodes)
     table.insert(lno_text, build_lno_str(line_num, gutter_width-1))
   end
 
+  side_scroll_contexts()
```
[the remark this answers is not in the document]
rwblokzijl: This is now only computed when horizontally scrolled

lua/treesitter-context.lua
```diff
@@ -315,6 +315,18 @@ local function get_gutter_width()
   return vim.fn.getwininfo(vim.api.nvim_get_current_win())[1].textoff
 end
 
+local window_scrolled_horizontal
+do
+  local leftcol
```
lewis6991, line 320: Doesn't this need to be specific to a window?

lua/treesitter-context.lua
```diff
@@ -749,11 +773,17 @@ end
...
+    update()
+    if window_scrolled_horizontal() then
+      horizontal_scroll_contexts()
+    end
+  end)
 
   autocmd('CursorMoved', function()
     if cursor_moved_vertical() then
       update()
+    else
```
lewis6991, line 786: Remove

lua/treesitter-context.lua
```diff
@@ -621,6 +633,18 @@ local function get_relative_line_num(ctx_node_line_num)
   return cursor_line_num - ctx_node_line_num - num_folded_lines
 end
 
+local function horizontal_scroll_contexts()
+  if context_winid == nil then
+    return
+  end
+  local current_win_view = vim.api.nvim_win_call(context_winid, function ()
+    return vim.fn.winsaveview()
```
lewis6991, line 641: I think this would be simpler if you just called `winsaveview` for the main window and the context window and see if they are different. If they are, you can then call `winrestview({leftcol = ...})`.

lua/treesitter-context.lua
rwblokzijl: True, but since there is only 1 context window open for 1 relevant window at a time the variable can be reused. Now that i think of it there is 1 edge case where 2 windows are scrolled the same amount. Then the horizontal scroll will be missed after window change. I will fix that.

lua/treesitter-context.lua
rwblokzijl: `winsaveview` returns more information than just `leftcol`. I can build in a check and only call `winrestview` on change, but since the function is now called on horizontal scroll only, this only avoids a single call every window change.

lewis6991: > winsaveview returns more information than just leftcol.

That shouldn't be a problem. `winsaveview()` is actually much cheaper than I thought as it just returns internal variables:

```c
/// "winsaveview()" function
static void f_winsaveview(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
{
  tv_dict_alloc_ret(rettv);
  dict_T *dict = rettv->vval.v_dict;

  tv_dict_add_nr(dict, S_LEN("lnum"), (varnumber_T)curwin->w_cursor.lnum);
  tv_dict_add_nr(dict, S_LEN("col"), (varnumber_T)curwin->w_cursor.col);
  tv_dict_add_nr(dict, S_LEN("coladd"), (varnumber_T)curwin->w_cursor.coladd);
  update_curswant();
  tv_dict_add_nr(dict, S_LEN("curswant"), (varnumber_T)curwin->w_curswant);

  tv_dict_add_nr(dict, S_LEN("topline"), (varnumber_T)curwin->w_topline);
  tv_dict_add_nr(dict, S_LEN("topfill"), (varnumber_T)curwin->w_topfill);
  tv_dict_add_nr(dict, S_LEN("leftcol"), (varnumber_T)curwin->w_leftcol);
  tv_dict_add_nr(dict, S_LEN("skipcol"), (varnumber_T)curwin->w_skipcol);
}
```

It is `winrestview()` which does a bunch of stuff.

rwblokzijl: If i only call `winrestview` if the scroll is different, should i move it back into the update function then? I feel like that keeps the code a whole lot simpler.

lewis6991: Ok

rwblokzijl: Done

lua/treesitter-context.lua
```diff
@@ -621,6 +621,19 @@ local function get_relative_line_num(ctx_node_line_num)
   return cursor_line_num - ctx_node_line_num - num_folded_lines
 end
 
+local function horizontal_scroll_contexts()
+  if context_winid == nil then
+    return
+  end
+  local active_win_view = vim.fn.winsaveview()
+  local context_win_view = vim.api.nvim_win_call(context_winid, function () return vim.fn.winsaveview() end)
```
lewis6991, line 629:
```suggestion
  local context_win_view = vim.api.nvim_win_call(context_winid, vim.fn.winsaveview)
```

lua/treesitter-context.lua
```diff
@@ -621,6 +621,19 @@ local function get_relative_line_num(ctx_node_line_num)
...
+  if active_win_view["leftcol"] ~= context_win_view["leftcol"] then
+    context_win_view["leftcol"] = active_win_view["leftcol"]
+    vim.api.nvim_win_call(context_winid, function ()
```
lewis6991, line 632:
```suggestion
    api.nvim_win_call(context_winid, function ()
```

lua/treesitter-context.lua
```diff
@@ -621,6 +621,19 @@ local function get_relative_line_num(ctx_node_line_num)
...
 end
 
+local function horizontal_scroll_contexts()
+  if context_winid == nil then
+    return
+  end
+  local active_win_view = vim.fn.winsaveview()
+  local context_win_view = vim.api.nvim_win_call(context_winid, function () return vim.fn.winsaveview() end)
+  if active_win_view["leftcol"] ~= context_win_view["leftcol"] then
+    context_win_view["leftcol"] = active_win_view["leftcol"]
```
lewis6991, line 631:
```suggestion
  if active_win_view.leftcol ~= context_win_view.leftcol then
    context_win_view.leftcol = active_win_view.leftcol
```

lua/treesitter-context.lua
```diff
@@ -621,6 +621,19 @@ local function get_relative_line_num(ctx_node_line_num)
...
+    vim.api.nvim_win_call(context_winid, function ()
+      return vim.fn.winrestview(context_win_view)
```
lewis6991, line 633:
```suggestion
      return vim.fn.winrestview({leftcol = context_win_view.leftcol})
```

lewis6991 (changes requested):
Some suggestions.

lewis6991:
Can you just confirm this works before we merge?

lewis6991 (approved):

rwblokzijl:
It works :)

lewis6991:
Awesome. Thanks a lot for this!

PR #208, opened 2023-03-07
Show context for `table_constructor` in lua

aaronkollasch:
Within lua files such as neovim config files, I find many table constructors longer than the window size, where a context line would be helpful. For example:

image

This PR adds a context line for lua table constructors by default.


aaronkollasch:
Actually, this is fixed in #198.

PR #210, opened 2023-03-08
Add terraform default type patterns

rwblokzijl:
Re adds the default patterns from https://github.com/nvim-treesitter/nvim-treesitter-context/pull/156.

Terraform uses `block`s for all their top level data structures. Ideally I'd add a query for those that have an identifier that starts with `resource`, `data`, `locals` or `provider`. But i need some advice on how to do this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ use 'nvim-treesitter/nvim-treesitter-context'
   - [x] `smali`
   - [x] `swift`
   - [x] `teal`
+  - [x] `terraform`
   - [x] `toml`
   - [x] `tsx`
   - [x] `typescript`
@@ -176,7 +177,6 @@ use 'nvim-treesitter/nvim-treesitter-context'
   - [ ] `svelte`
   - [ ] `sxhkdrc`
   - [ ] `t32`
-  - [ ] `terraform`
   - [ ] `thrift`
   - [ ] `tiger`
   - [ ] `tlaplus`
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -93,6 +93,8 @@ local is_valid = cache.memoize(function(node, query)
   range[3] = range[1]
   range[4] = -1
 
+  local node_srow = range[1]
+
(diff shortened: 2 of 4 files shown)


lewis6991:
Awesome. Can you update the README?

rwblokzijl:
I will update the docs, but i would like to add one more query and i need some advice:

I would like to add a query where as long as the cursor is inside `for_expr` to show the first line of the `for_intro` 

image

lewis6991:
Screenshots are really unhelpful and I don't know this language. Can you share some code and what the file extension is.

lewis6991:
Ok I've had a look and I don't think this is possible due to the terraform grammar.

For whatever reason `for_object_expr` starts on the `{`. Additionally, `for_intro` is not a parent node, but instead adjacent to a parent node.

This is the best I could do:

```
(for_object_expr
  (_)
  (for_intro)
  (_) @context.end
) @context
```

Which gives the context:

```terraform
input_deps = { for slug, group in local.t2r_map :
```

lewis6991:
Ok, I've added a new capture `@context.start` to support AST structures like this.

rwblokzijl:
@lewis6991 thanks so much for looking into this.

However it does seem that `(attribute)` in `terraform.scm` does not work when the `for_object_expr` is set.
Notice that `input_deps` is not shown:

image
 for 
```scm
(for_object_expr
  (_)
  (for_intro) @context.start
  (_) @context.end
) @context

([
  (block)
  (object)
  (object_elem)
  (attribute)
  (template_literal)
] @context)
```
But it is shown:
image

 for 
```scm
; (for_object_expr
;  (_)
;   (for_intro) @context.start
;   (_) @context.end
; ) @context

([
  (block)
  (object)
  (object_elem)
  (attribute)
  (template_literal)
] @context)
```

I think this is because `for_object_expr` starts on the same line as the `attribute`

lewis6991:
We are always going to be limited by the grammar structure, and it seems the terraform grammar makes this too difficult.

PR #256, opened 2023-04-29
graphql support

uchars:
not really sure if this is needed, but if 'foo' in 
```graphql
foo {
  foo1,
  foo2,
  foo3,
  ...
}
```
should also be highlighted 

```(arguments)``` would have to be removed from ```(selection( field```


uchars:
wups, looks like i made a checkout oopsie

PR #291, opened 2023-06-17
Add support for haskell

bartoszluka:
Tested locally and it works great. Not sure if there should be more nodes that the `context` should capture but they can be added in the future in the PR.

The image shows that how it look when running in the Haskell test file:

plugin working for haskell file

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ use 'nvim-treesitter/nvim-treesitter-context'
   - [x] `glimmer`
   - [x] `go`
   - [x] `graphql`
+  - [x] `haskell`
   - [x] `html_tags`
   - [x] `ini`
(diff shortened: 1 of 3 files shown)


vsoftco:
@bartoszluka Starting with nvim 0.10.0, the query is not valid anymore. I am getting 

```
Error detected while processing CursorMoved Autocommands for "*":
Unable to load context query for haskell:
...m/0.10.0/share/nvim/runtime/lua/vim/treesitter/query.lua:252: Query
 error at 2:4. Invalid node type "alt":
```

The problem are the `(alt)` and `(exp_case)` nodes in `context.scm`.

bartoszluka:
I will look into it in the evening

PR #372, opened 2023-12-15
feat(dart): multiline support

genesistms:


genesistms:
I Couldn't find solution to missing `else` in `(if_statement)`. Help to build a query would be nice. Tree looks like this:

genesistms:
Note for _future anyone_ the reason we cannot properly show multiline `(function_signature)` like:

```dart
void test(
  int a,
  int b,
) {
  // ...
}
```
is that we cannot do sibling group in neovim yet:
```query
(
  (function_signature)
  (function_body) @context.end
) @context
```

PR #389, opened 2024-02-07
feat: Add perl support

mbudde:
Try-catch support is currently limited because the treesitter tree is missing some nodes for the catch and finally parts.


mbudde:
Oh, did not see there's already an open PR for Perl support 😬

Issue #410, opened 2024-02-27
Very poor performance after “do not use vim.treesitter.highlighter”


amarshall:
### Description

I am seeing huge performance regressions on some files (usually large ones or those with a lot of nesting) on current master (b8d1ffe58a88e0356da56b167373e89c4579ce15). I have bisected this to 4f5c7426978043b8eb76cf1cde44f399cf09ea0b, before that change (or reverting it) the problem goes away.

### Neovim version

v0.9.5

### Expected behavior

No lag upon context change or appearance

### Actual behavior

Intolerable lag

### Minimal config

```Lua
local plugins = {
  ts         = 'https://github.com/nvim-treesitter/nvim-treesitter',
  ts_context = 'https://github.com/nvim-treesitter/nvim-treesitter-context',
  -- ADD ADDITIONAL PLUGINS THAT ARE _NECESSARY_ TO REPRODUCE THE ISSUE
}

for name, url in pairs(plugins) do
  local install_path = '/tmp/nvim/site/'..name
  if vim.fn.isdirectory(install_path) == 0 then
    vim.fn.system { 'git', 'clone', '--depth=1', url, install_path }
  end
  vim.o.runtimepath = install_path..','..vim.o.runtimepath
end

-- ADD INIT.LUA SETTINGS THAT IS _NECESSARY_ FOR REPRODUCING THE ISSUE
```

### Steps to reproduce

1. `nvim --clean -u minimal.lua` with some file (I used [this file](https://github.com/NixOS/nixpkgs/blob/73de017ef2d18a04ac4bfd0c02650007ccb31c2a/pkgs/top-level/all-packages.nix))
2. `:TSContextEnable`
3. Press `M` to move cursor to the middle (so context appears)
4. Scroll enough to get some context
5. Press `H` to move the cursor to the top (so context disappears)
6. Press `M` and experience lag for cursor movement and context reappearing (does not occur with `:TSContextDisable`)

Also occurs when scrolling through and the context is visible and needs to change.

colin969:
The lag is easily half a second plus for a few thousand line file.

Definitely a deal-breaker, hopefully it'll be a relatively quick fix.

romgrk:
I have reverted the commit for now.

PR #414, opened 2024-03-05
Perf: revert "fix: do not use vim.treesitter.highlighter"

romgrk:
This reverts commit 4f5c7426978043b8eb76cf1cde44f399cf09ea0b, following the reports in #410.


romgrk:
I'm unable to reproduce the issue in #387 (highlight) and it looks much less harmful than the current performance issue, so I'll merge the revert as is. Let's figure out a solution for the other one in a follow-up.

PR #419, opened 2024-03-19
add multi window support

xzbdmw:
This will track a global map that map winid to {bufnr,gut_bufnr, context_bufnr,gutter_winid,context_winid} and dynamically clean up and add new one, I'm new to lua so thanks for your advice. Currently WinScroll autocmd seems don't adjust context window position and I can't figure out why.


xzbdmw:
Oh why the formatting is changed

lua/treesitter-context/render.lua
```diff
@@ -7,48 +7,49 @@ local config = require('treesitter-context.config')
...
-  return gutter_bufnr, context_bufnr
-end
+-- --- @return integer gutter_bufnr
+-- --- @return integer context_bufnr
+-- local function get_bufs()
+--   context_bufnr = create_buf(context_bufnr)
+--   gutter_bufnr = create_buf(gutter_bufnr)
+--
+--   return gutter_bufnr, context_bufnr
+-- end
```
romgrk, line 38: Don't leave commented code, either use it or remove it.

lua/treesitter-context/render.lua
```diff
@@ -7,48 +7,49 @@ local config = require('treesitter-context.config')
...
-local context_winid --- @type integer?
-
---- @param buf integer?
---- @return integer buf
-local function create_buf(buf)
-  if buf and api.nvim_buf_is_valid(buf) then
-    return buf
+--- @param winid integer?
+--- @return {bufnr:integer,gut_bufnr: integer, context_bufnr: integer,gutter_winid:integer?,context_winid:integer?} buf
+local function create_buf(winid)
```
romgrk, line 14: Doesn't return just one buffer, name should probably evolve.

lua/treesitter-context/render.lua
```diff
@@ -7,48 +7,49 @@ local config = require('treesitter-context.config')
 local ns = api.nvim_create_namespace('nvim-treesitter-context')
 
 -- Don't access directly, use get_bufs()
-local gutter_bufnr --- @type integer?
-local context_bufnr --- @type integer?
+local bufnr_maps = {} --- @type table<integer, {bufnr:integer?,gutter_bufnr: integer?, context_bufnr: integer?,gutter_winid:integer?,context_winid:integer?}>
```
romgrk, line 10: Name isn't ideal, it's not just a `bufnr` map, it also as window ids.

romgrk (commented):
I haven't tested yet but overall the approach & feature looks good to me.

lua/treesitter-context.lua
```diff
@@ -132,7 +127,13 @@ function M.enable()
...
   attached[cbuf] = true
 
-  autocmd({ 'WinScrolled', 'BufEnter', 'WinEnter', 'VimResized' }, update)
+  autocmd({ 'BufEnter', 'WinEnter', 'VimResized' }, update)
+
+  autocmd({ 'WinScrolled' }, function()
+    for winid, bufnr_map in pairs(require('treesitter-context.render').get_bufnr_maps()) do
+      update(bufnr_map.bufnr, winid)
+    end
+  end)
```
romgrk, line 136: Are we updating all the windows on scroll? Shouldn't we be only updating the scrolled one?

xzbdmw: Commands such as split will trigger winscroll, and every window width should be adjusted to new width, how should we differentiate scroll and resizing? Should I use winresized instead?

xzbdmw:
There is one thing I can't understand
in neovide, resizing works perfect:

https://github.com/nvim-treesitter/nvim-treesitter-context/assets/97848247/3631c3e0-a9eb-4d97-ab2a-7032249d200c

but in kitty, when new window created, the position becomes wrong:

https://github.com/nvim-treesitter/nvim-treesitter-context/assets/97848247/44f4967d-f10b-49e2-ae08-566a8cfd8686

Can you help with me what causes the difference?

xzbdmw:
@romgrk @lewis6991 Hello could you test this? The only issue remains is the attached[buf] I'm not sure how to change it, since now we have multiple windows.

romgrk:
For your issue, I don't like adding an async delay there, even if just `1ms`. You should retrieve the buf/win ID during the event handler via the usual mechanism (`<amatch>` and friends) and pass it as an argument to `update()`, maybe an optional param would be good.

xzbdmw:
> For your issue, I don't like adding an async delay there, even if just `1ms`. You should retrieve the buf/win ID during the event handler via the usual mechanism (`<amatch>` and friends) and pass it as an argument to `update()`, maybe an optional param would be good. Though I don't see the `defer_fn` in this PR, lmk when it's ready for review.

vim.schedule works fine and simpler, and I leave all attached[buf] unchanged, should be ready for review and testing:)

lua/treesitter-context.lua
```diff
@@ -132,7 +143,39 @@ function M.enable()
...
+
+  autocmd({ 'WinResized' }, function()
+    local event = vim.api.nvim_get_vvar('event')
+    local window_ids = event.windows
+    for stored_winid, window_context in
+      pairs(require('treesitter-context.render').get_window_contexts())
+    do
+      for _, window_id in pairs(window_ids) do
+        if stored_winid == window_id then
+          throttle(function()
```
lewis6991, line 159: This doesn't work for anonymous functions since it is only ever called once.

lua/treesitter-context/render.lua
```diff
@@ -6,49 +6,71 @@ local config = require('treesitter-context.config')
...
+WindowContext.__index = WindowContext
 
-  vim.bo[buf].undolevels = -1
-  vim.bo[buf].bufhidden = 'wipe'
+local windowContexts = {} --- @type table<integer, WindowContext>
 
-  return buf
+--- @param winid integer?
+local function win_close(winid)
+  vim.schedule(function()
```
lewis6991, line 22: The code doesn't show why this is needed.

xzbdmw: I move win_close to top since store_context needs to call close if underline buffer changed

lua/treesitter-context.lua
xzbdmw: Do you mean throttle will collapse multiple calls into one call, if so, I should directly call the anonymous functions?

lewis6991: Calling throttle like this (`throttle(...)()`) doesn't work. It needs to be.

```lua
local f = throttle(...)
f()
f() -- throttled
```

```lua
throttle(...)()
throttle(...)() -- not throttled!
```

lua/treesitter-context.lua
```diff
@@ -155,7 +198,18 @@ function M.enable()
 
   autocmd({ 'BufLeave', 'WinLeave' }, close)
```
xzbdmw, line 199: I think maybe bufleave and winleave is no longer needed, winleave can be replaced by winclosed, and bufleave's close behavior is expressed in store_contest's win_close call.

lua/treesitter-context/render.lua
```diff
@@ -60,21 +82,21 @@ local function display_window(bufnr, winid, width, height, col, ty, hl)
...
       height = height,
       row = 0,
       col = col,
     })
   end
-  return winid
+  return float_winid
 end
 
 --- @param winid integer
```
xzbdmw, line 102: textoff is inconsistent too. In my case it will return 4 and after next update time it will return 8, cause window position to flash

xzbdmw:
@romgrk hello do you have time to review?
According to this suggestion
https://www.reddit.com/r/neovim/comments/10eu7go/comment/j4uaxpe/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button
I'm wrapping WinEnter in a vim.schedule now to get new bufnr. It should be ready to merge, after 3 weeks of use I see no problem.

romgrk (approved):
LGTM, I'll run the PR in my config to test it.

lua/treesitter-context.lua
romgrk, line 1: > I'm wrapping WinEnter in a vim.schedule now to get new bufnr. It should be ready to merge, after 3 weeks of use I see no problem.
But I have not find a reliable way to get buffer information after WinEnter is triggered, the solution above is not perfect if you open many windows at the same time, since vim.schedule's execute time is uncertain, as the origin comment says. But it is still a improvement since there would not be some file buffer unrelated windows to show previous buf's context.

Yeah sometimes events are ugly. Maybe this would be better fixed in neovim directly, if it's unreliable.

xzbdmw: Hi, I remove WinEnter since it is not reliable, and indeed there are no need for that autocmd

lua/treesitter-context/render.lua
```diff
@@ -6,49 +6,71 @@ local config = require('treesitter-context.config')
...
+--- @field gutter_bufnr integer The gutter buffer number
+--- @field context_bufnr integer The context buffer number
+--- @field gutter_winid integer? The window ID of the gutter
+--- @field context_winid integer? The window ID of the context
+local WindowContext = {}
+WindowContext.__index = WindowContext
 
-  vim.bo[buf].undolevels = -1
-  vim.bo[buf].bufhidden = 'wipe'
+local windowContexts = {} --- @type table<integer, WindowContext>
```
romgrk, line 18: `window_contexts`, consistent naming with the rest.

xzbdmw: done

lua/treesitter-context.lua
romgrk: Wdym by not needed? I'm observing a few cases where the context doesn't show up, wondering if it's connected?

lua/treesitter-context.lua
```diff
@@ -153,9 +191,18 @@ function M.enable()
...
 
-  autocmd({ 'BufLeave', 'WinLeave' }, close)
+  autocmd({ 'WinClosed' }, function(args)
+    local winid = tonumber(args.match)
+    for stored_winid, _ in pairs(require('treesitter-context.render').get_window_contexts()) do
+      if winid == stored_winid then
+        close(stored_winid)
+      end
+    end
+  end)
```
romgrk, line 201: Could we keep the same pattern as other events, and assign the function to a name and pass a ref to it here, instead of defining it literally here? Keeps things more ordered and helps reify what's happening.

lua/treesitter-context.lua
```diff
@@ -153,9 +191,18 @@ function M.enable()
...
 
-  autocmd('User', close, { pattern = 'SessionSavePre' })
+  autocmd('User', function()
+    close_all()
+  end, { pattern = 'SessionSavePre' })
```
romgrk, line 205: `function() close_all() end` is basically equivalent to just `close_all` here.

lua/treesitter-context.lua
xzbdmw: > Have you observed similar behavior

No, but I have context window opened in none-file buffers if I just keep WinEnter unchanged as is.
截屏2024-04-19 21 08 03
I have a keymap that first enter right "reference list", then immediately move focus to the preview window, and a workaround is wrapping callback in vim.schedule, If you prefer keep WinEnter I'll add it back.

romgrk: I don't have a preference, whatever works is best, I'm just wondering if changing the events might have caused the few cases I saw where the window wasn't appearing. I'm always very confused with autocmds so I couldn't tell you which combination of BufEnter/WinEnter/BufWinEnter is right for our use-case.

xzbdmw:
@romgrk @lewis6991 Hello any chance this get merged?

xzbdmw:
Hmm no problem for me so far, works seamless

apollo1321:
Could you please clarify the reason this PR was closed?

I have developed my own implementation of multi-window support a couple of months ago, and I have been using it extensively since then. My experience has been very positive; it has performed flawlessly without any issues related to multiple windows or resizing.

Additionally, I have conducted performance testing using flamegraphs, and I have observed that treesitter-context consumes approximately 7% of processing time, which appears to be within a reasonable range.

I believe that multi-window support is a highly valuable feature, especially for working with large codebases efficiently. I think it would greatly benefit other users as well.

If it would be helpful, I am more than happy to share my implementation and potentially create a new PR.

xzbdmw:
Because I tried to rebase but there are too many commits behind 😄

romgrk:
Not enough time to review/debug the few minor issues that appeared at the time. I'd be happy to see this feature, I just don't have much time to put into FOSS right now, and @lewis6991 also needs to sign off on it as he's the more active maintainer at the moment. But please open a PR if you have one that's production ready.

PR #473, opened 2024-07-13
docs: move vimdoc to `doc/`

ss-raicangu:
First off, thanks for writing a vimdoc :) . Apologies for not creating an issue to consult first. Thought a PR would be faster.

* Problem: `:helptags` don't seem to be read when they are generated at the root of the project.
* Solution: `:help 'runtimepath'` suggests that Neovim looks in the `doc/` directory for documentation tags. Moving the vimdoc file to `doc/` and generating `:helptags` seems to work. Tested with `:help nvim-treesitter-context`.


ss-raicangu:
Thanks for considering/merging 🙂 .

PR #505, opened 2024-10-17
fix(render): add missing window check - follow ups

przepompownia:


przepompownia:
I'm not sure if the check should be performed here. 

Notice another unchecked `api.nvim_win_get_buf` 17 lines below this change. I haven't caught the invalid window error there yet.

przepompownia:
Browsing conflicts with Diffview can expose such errors.

przepompownia:
[this place](https://github.com/nvim-treesitter/nvim-treesitter-context/blob/6cba006378b0ca68a66ea54153c0113724c91753/lua/treesitter-context/render.lua#L415) still waits for catch…

apollo1321:
> https://github.com/nvim-treesitter/nvim-treesitter-context/blob/6cba006378b0ca68a66ea54153c0113724c91753/lua/treesitter-context/render.lua#L415
> 
> This place still waits for catch…

How is it possible that the context window is valid, but its buffer has somehow become invalid?

przepompownia:
> How is it possible that the context window is valid, but its buffer has somehow become invalid?

Right, I didn't notice that the value mentioned earlier as uncaught is validated in the second commit :grinning:

PR #508, opened 2024-10-18
fix: multiwindow support; fix bugs

apollo1321:
https://github.com/nvim-treesitter/nvim-treesitter-context/issues/507

### DIFF
--- a/lua/treesitter-context.lua
+++ b/lua/treesitter-context.lua
@@ -39,8 +39,14 @@ end
 
 local attached = {} --- @type table<integer,true>
 
-local function close()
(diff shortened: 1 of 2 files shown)


lua/treesitter-context.lua
```diff
@@ -61,15 +67,14 @@ local function cannot_open(bufnr, winid)
     or vim.bo[bufnr].filetype == ''
     or vim.bo[bufnr].buftype ~= ''
     or vim.wo[winid].previewwindow
-    or vim.fn.getcmdtype() ~= ''
```
apollo1321, line 64: Deleting random window when in cmdline causes window context flickering.

PR #556, opened 2025-02-02
fix(lua): Filter comment-only lines from context view

### LINKED ISSUE #476: Filter blank and comment-only lines from context view
It would be great if this tool could filter blank and comment-only lines from the context view.

At present, with `multiline_threshold = 3`, I see this:

image

It would be great if the tool could either filter this down the single relevant line of context (`if vim.g.neovide then`), or use this as an opportunity to display a second relevant line that is currently not shown, like this:

```
callback = function()
  if vim.g.neovide then
```

Crypto-Spartan:
fixes #476 (only for Lua)

### DIFF
--- a/queries/lua/context.scm
+++ b/queries/lua/context.scm
@@ -11,7 +11,8 @@
 ) @context
 
 (function_definition
-  body: (_) @context.end
+  parameters: (_)
+    _ @context.end
 ) @context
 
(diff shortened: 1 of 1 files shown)


Crypto-Spartan:
I've been using this in my config since September with no issues

PR #583, opened 2025-04-12
test: properly check treesitter lang and only install parser when needed

PeterCardenas:
Because of the way we were checking the treesitter lang, we were relying
on the parser to already be installed for the test to not be skipped.
Now we check it via an `nvim-treesitter` API that does not require the
parser to be installed. This results in all the parsers to be installed
even when the related tests won't be run, so we do a lazy setup here as
well.

Followup to #568

### DIFF
--- a/test/contexts_spec.lua
+++ b/test/contexts_spec.lua
@@ -77,11 +77,7 @@ setup(function()
     --- @type string
     local treesitter_lang = exec_lua(
       [[
-    local ok, parser = pcall(vim.treesitter.get_parser, ...)
-    if not ok then
-      return nil
-    end
-    return parser:lang()
+    return require('nvim-treesitter.parsers').get_buf_lang(...)
     ]],
       bufnr
     )
@@ -111,7 +107,7 @@ for _, lang in ipairs(langs_with_queries) do
         return
       end
 
-      setup(function()
+      lazy_setup(function()
         cmd([[let $XDG_CACHE_HOME='scratch/cache']])
         install_langs(lang)
       end)


test/contexts_spec.lua
```diff
@@ -77,11 +77,7 @@ setup(function()
     --- @type string
     local treesitter_lang = exec_lua(
       [[
-    local ok, parser = pcall(vim.treesitter.get_parser, ...)
-    if not ok then
-      return nil
-    end
-    return parser:lang()
+    return require('nvim-treesitter.parsers').get_buf_lang(...)
```
lewis6991, line 80: `nvim-treesitter` is going to be replaced with the `main` branch soon enough, so I'd like to minimise relying on any it's API.

PeterCardenas:
so right now if you run a test with no populated `deps` directory, it'll skip the tests pretending that they passed. this is particularly useful for the `make test FILTER=test.<ft>` workflow where you just end up installing the parsers you need for the test you write

test/contexts_spec.lua
PeterCardenas: so looks like in main we register the filetype to lang mappings [immediately](https://sourcegraph.com/github.com/nvim-treesitter/nvim-treesitter@873e118e1228d43540e076024710bc873fe9b74f/-/blob/plugin/filetypes.lua?L64) instead of [lazily](https://github.com/nvim-treesitter/nvim-treesitter/blob/7bbed4b548d9a81c14351bda93d16336edefd067/lua/nvim-treesitter/parsers.lua#L59-L68) in which case we can use `vim.treesitter.language.get_lang` directly. we could change it to that anticipatorily, just that this PR won't work until main is swapped in

lewis6991: I think we can safely switch to `main` in this repo even though it isn't default yet.

PeterCardenas: looks like the `main` branch is using an incompatible `vim.validate` signature than the one the runner is using. the runner expects a table, but we're giving a string

lewis6991: We can update the runner to 0.11.

lewis6991: We only have 1 runner version and several test versions. The runner version should ideally be the latest release.

PR #591, opened 2025-05-02
refactor: expose context window

phanen:
## Problem
To attach `treesitter-context` on a float window, we need to create context with a larger `zindex` than the float window.

However there's no way to properly achieve it, since all context win use the same `config.zindex`.
(so larger `config.zindex` cause float window overlap with the context attached on non-float window)

image

## Solution
~Make `render.open` accept an `win_config`. (But I'm not sure if this will increase maintain burden here).~

Another way is expose `context_winid`, so we can fix `zindex` later by `nvim_win_set_config`.

(A downstream fix https://github.com/ibhagwan/fzf-lua/pull/2023 via this)

### DIFF
--- a/lua/treesitter-context/render.lua
+++ b/lua/treesitter-context/render.lua
@@ -481,6 +481,7 @@ end
 --- @param winid integer
 --- @param ctx_ranges Range4[]
 --- @param ctx_lines string[]
+--- @return WindowContext?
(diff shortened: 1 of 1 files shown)


phanen:
Close as it can be got via `debug.getupvalue`.

PR #603, opened 2025-05-25
fix(enhancement): Fixes #453 - support nearest parent

### LINKED ISSUE #453: Allow jumping to nearest context with go_to_context()
I wanted to have a mapping to jump to whatever is the current "parent" node, and found the `go_to_context()` function which works great! But it only jumps to contexts that are also displayed by the plugin, and I'd like this mapping to be more granular and also consider contexts that are in view, or hidden by my `max_lines` setting.

I managed to hack something together if I expose the `iter_context_parents` function from the plugin:

```lua
-- add in `lua/treesitter-context/context.lua`
M.iter_context_parents = iter_context_parents

-- in my mapping:
local context = require('treesitter-context.context')
local cursor = vim.api.nvim_win_get_cursor(0)
local row = cursor[1]
local parents, _query = context.iter_context_parents(0, { row - 1, 0, row - 1, 0 })()

for i = 1, #parents do
  local parent_row = parents[i]:range() + 1
  if parent_row < row then
    vim.cmd([[ normal! m' ]])
    vim.api.nvim_win_set_cursor(0, { parent_row, 0 })
    vim.cmd.normal('^')
    return
  end
end
```

This seems to work as I want, and now I wonder what it would take to implement a cleaner version of this inside the plugin, or what bits could be exposed to allow a custom mapping like this (`iter_context_parents` seems a bit too low-level).

Also not sure if there's a way to do this using e.g. `nvim-treesitter-textobjects`, it seems the magic sauce that's needed is the queries collected in this plugin.

isakbm:


lua/treesitter-context/context.lua
```diff
@@ -366,6 +368,15 @@ function M.get(bufnr, winid)
               context_lines[#context_lines + 1] = lines
             end
           end
+        else
+          local range0 = context_range(parent, bufnr, query)
+          if range0 and range_is_valid(range0) then
+            local range, _lines = get_text_for_range(range0, bufnr)
+            -- TODO: consider printing lines here for debug purposes
```
lewis6991, line 375: Remove please

lewis6991 (commented):
Commit message should be prefixed with `feat:`. And like the last PR, please put issue references in the commit message summary on a separate line (see [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue))

I think we should remove the old behaviour and make this the new behaviour. Whether the context is being displayed in the context window or not shouldn't influence navigation like this.

lua/treesitter-context/context.lua
isakbm: Some people might want to keep the old behavior too. This way it becomes easy and natural to do so, as the ranges are split into

- context window ranges
- visible window ranges

That being said, perhaps I misunderstood your point? 🤔

isakbm: Perhaps what you mean is that I should refactor the logic a bit so that the way to avoid some a mount of repetition there between the two paths of the `if` - `else` block?

isakbm: I tried to refactor it a bit but that just breaks it, and also doesn't add up or make sense to me. So I think it needs to be kept the way it is now. The repetition is very minimal anyway, and quite straight forward and local.

lewis6991: I meant to simply remove the TODO comment.

isakbm: alright I've now removed this TODO comment 🤞

lua/treesitter-context.lua
```diff
@@ -310,14 +315,23 @@ function M.setup(options)
   end
 end
 
+--- jump to the nearest context parent
+---
 --- @param depth integer? default 1
-function M.go_to_context(depth)
+--- @param mode? 'context window' | 'entire window' default 'context window', the parents to pick from when jumping
```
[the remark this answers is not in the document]
isakbm: It is easy for me to make this change, and if this is required for you in order to accept the PR, then I am happy to do so.

Just a fair warning though, this might be unexpected to other users who are already used to the current behavior.

Also, it could be nice to be able to configure this, because then people can make two separate key-bindings like

```lua
vim.keymap.set("n", "[C", function()
  require("treesitter-context").go_to_context(vim.v.count1, "context window")
end, { silent = true })

vim.keymap.set("n", "[c", function()
  require("treesitter-context").go_to_context(vim.v.count1, "entire window")
end, { silent = true })
```

This way `[C` would jump to first parent in the context window, while `[c` would jump to first parent nearest the cursor.

Let me know if you're still certain about this hard coded change of behavior. If you are then like I said I'll happily make this change.

I might keep this opt in configurable feature in my own fork though.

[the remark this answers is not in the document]
lewis6991: Yes I'm well aware.

When reviewing this, it really isn't clear at all how the two different modes differ and the names are very obscure. That's a clear signal to me that this shouldn't be configurable.

There are near infinite ways we can make this plugin configurable, but I'm not interested in maintaining such a project, and I'm not interested in appeasing everyone. I wasn't keen on adding the feature of navigation to begin with but relented when there was enough desire.

And just to be clear, we don't simply implement every feature any person that raises an issue asks for, so please don't continue to raise PR's for every feature request. I'm mostly interested in maintaining a robust, well implemented, well tested plugin with as little configuration as possible. That's there if spent most of my time working on this, particularly the performance and quality of the highlights.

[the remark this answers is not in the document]
lewis6991: For now I've made up my mind. People are free to raise issues if they feel strongly about the old behaviour. Though as mentioned it's not usual for vim motions to be predicated on whether it is visible in the viewport or not, imagine if the `n`  key behaved like that.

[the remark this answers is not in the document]
isakbm: also the naming no longer makes sense, because it is not `go_to_context` (window), it is now, go to nearest parent 🤔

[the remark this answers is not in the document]
lewis6991: I'm now contemplating if we need to support this new feature at all.

Implementing something like `go_to_parent` is very simple without this plugin. You just use `vim.treesitter.get_node():parent():range()` and pass that to `nvim_win_set_cursor()`. For range you just call `parent()` in a loop.

Maybe we should abandon this and just document how to implement go to parent?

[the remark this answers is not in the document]
isakbm: Yeah but it does seem nice to unify it somewhat, or rather to package it in with this plugin. Like what if in the README and the docs like you suggest, we provide some keymaps or something. Example usage. I think it would also be cool btw to be able to jump to not just the nearest parent in the `context window` but also the `farthest` parent in the context window. Now these two jumps are something that should of course be done by `treesitter-context` and not by the end user.

Maybe we need to come up with a more general approach. I think it would be quite cute and simple to just have the following methods

`go_to_context_parent(n)`

where if `n <= 0` it goes to the farthest (top most parent), and if n = 1, 2, 3, ... it goes successively deeper into the context.

then to get the behavior that we now made as default behavior, we would instead just leave that up to the end user.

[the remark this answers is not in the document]
isakbm: I find the code change in the `treesitter-context` to be so minimal, that I would love to just have it there. Honestly if this is going to keep on going like this, I just don't have time for a further discussion, and would rather just live with my own fork that I'll periodically keep up to date with `origin` . I'm not in the business of including yet another plugin just for the sake of this one single feature. ^  ^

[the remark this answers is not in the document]
isakbm: My final enthusiastic proposal would be

- go_to_context(depth, n = 0)
- go_to_parent(depth, n = 0)

here n is intuitively similar behavior as max_lines , that is n = 0 goes to the deepest parent, and n = 1 , 2, 3, etc go to the highest parents successively

this would make me a very happy camper, and I think I need this so much that regardless I'll just have to have this in my own fork if it is not something that wants to be upstreamed.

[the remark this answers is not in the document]
isakbm: Also note that there is no distinction now between visual or not @lewis6991. It is either restricted to the `context window` or it is just `all the parents`.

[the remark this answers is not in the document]
isakbm: Also I've made it go to start of the line now @toupeira  with a call to 

```lua
vim.cmd([[normal! ^]])
```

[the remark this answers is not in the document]
isakbm: I think it's quite elegant and generic now

```lua
--- @param depth integer
--- @param contexts Range4[]
function go_to(depth, contexts)
  local line = api.nvim_win_get_cursor(0)[1]
  local context = nil

  for idx = #contexts, 1, -1 do
    local c = contexts[idx]
    if depth == 0 then
      break
    end
    if c[1] + 1 < line then
      context = c
      depth = depth - 1
    end
  end

  if context == nil then
    return
  end

  vim.cmd([[normal! m']]) -- add current cursor position to the jump list
  api.nvim_win_set_cursor(0, { context[1] + 1, context[2] })
  vim.cmd([[normal! ^]]) -- place cursor at start of text
end
```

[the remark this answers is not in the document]
toupeira: @isakbm IMHO that all sounds great and sensible, thanks for your work on this!

lua/treesitter-context.lua
```diff
@@ -310,31 +315,59 @@ function M.setup(options)
...
-  if context == nil then
+  if range == nil then
     return
   end
 
-  vim.cmd([[ normal! m' ]]) -- add current cursor position to the jump list
-  api.nvim_win_set_cursor(0, { context[1] + 1, context[2] })
+  vim.cmd([[normal! m']]) -- add current cursor position to the jump list
+  api.nvim_win_set_cursor(0, { range[1] + 1, range[2] })
+  vim.cmd([[normal! ^]]) -- place cursor at start of text
```
lewis6991, line 341: Line 340 is already setting the cursor to the correct place so this line shouldn't be needed.

isakbm: Yeah I was also expecting that to be the case, but observation would say otherwise. I can try to dig into this and figure out why it fails, but is it really that big of a deal to have a potentially redundant extra command here that does what needs to be done? We're already issuing a command to set the jump list mark. with `m'` anyway.

isakbm: Wait a second, I think I see something ...

lewis6991: `vim.cmd([[normal! ^]])` is not what we want here. We need to place the cursor at the start of the context, wherever that may be.

isakbm: then that's where we need to introduce a fix :)

isakbm: should I include a fix for this in this PR?

isakbm: I've now rebased on your latest refactor, and fixed a couple more bugs present on master

- too many nodes are included, last one is repeated, I simply remove it for now but maybe you have a better idea to fix underlying issue
- the column offset is now correctly set and not hard coded to zero
- rest of the feature improvements are still there

🤞 

@lewis6991

isakbm: I see that some tests are failing, I think I know why, but it's curious then why is sometimes an extra node added and other times not 🤔

isakbm: It's all fixed now, would be lovely if it could get merged in :)

toupeira: Gave it a quick try, seems to work great including moving to the correct column. Nice work! :+1:

isakbm:
Any chance that this can get merged in @lewis6991 ? 

I've simplified it a lot, and avoided that extra list of ranges 🤞

PR #604, opened 2025-05-27
fix: do not show context when within window

PeterCardenas:
Having an issue where the context shows up when the first line of the context is already in the window. This results in a case where I open a buffer and the second line is obfuscated by the separator. This fixes it in a somewhat hacky way by doing a check of the context lines to see if they are all already on the screen. I thought for a bit if I should try to fix this somewhere else, but there are a lot of edge cases and this seemed easiest/less risky.

### DIFF
--- a/lua/treesitter-context/context.lua
+++ b/lua/treesitter-context/context.lua
@@ -396,6 +396,14 @@ function M.get(bufnr, winid)
     end
   end
 
+  -- Do not show the context window if all contexts are below the top row.
+  local all_contexts_below_top_row =
+    not vim.tbl_contains(context_ranges, function(range) --- @param range Range4
+      return range[1] < top_row
+    end, { predicate = true })
+  if all_contexts_below_top_row then
+    return
+  end
   local trim = contexts_height - max_lines
   if trim > 0 then
     trim_contexts(context_ranges, context_lines, trim, config.trim_scope == 'outer')


isakbm:
FYI I think I know a fix for this, it's to only use the `separator` offset if the context contains more than 1 line ... seems kinda obvious now in hindsight @PeterCardenas

isakbm:
Yes here's the fix @PeterCardenas 

```diff
M lua/treesitter-context/context.lua
@@ -363,10 +363,13 @@ function M.get(bufnr, winid)
       for _, parent in ipairs(parents) do
         local parent_start_row = parent:range()
 
+        local num_context_lines = math.min(max_lines, contexts_height)
+
         -- NOTE: this avoids covering up context by separator line
-        local separator_offset = config.separator and 1 or 0
+        --       but is only needed when context is nontrivial of course
+        local separator_offset = (num_context_lines > 0 and config.separator) and 1 or 0
 
-        local contexts_end_row = top_row + separator_offset + math.min(max_lines, contexts_height)
+        local contexts_end_row = top_row + separator_offset + num_context_lines
 
         -- context_ranges: Only process the parent if it is not in view.
         if parent_start_row < contexts_end_row then
```

isakbm:
https://github.com/nvim-treesitter/nvim-treesitter-context/pull/605

PeterCardenas:
@isakbm that diff worked like a charm, closing this PR.

